PR #1, opened 2019-10-07
Make Platform Support Guarantees Explicit

mossheim:
RFC text: https://github.com/brianlheim/rfcs/blob/plfrm-sprt/rfcs/0001-make-platform-support-guarantees-explicit.md

### DIFF
--- a/rfcs/0001-make-platform-support-guarantees-explicit.md
+++ b/rfcs/0001-make-platform-support-guarantees-explicit.md
@@ -57,7 +57,7 @@ possible.
 The official minor release of SuperCollider is guaranteed to be compatible with:
 - major releases of Xcode (AppleClang), starting from 2 years prior to the minor release date of SuperCollider
 - major releases of gcc, MSVC, and clang, 4 years
-- minor releases of Qt, 3 years
+- minor releases of Qt, 2 years
 
 Other libraries and tools which SuperCollider uses, but which do not have such guarantees, are:
 - Boost (see note below)
@@ -147,7 +147,7 @@ would be:
 - MSVC: 2017 (2017-03) - now on 2019
 - gcc: 6 (2017-04) - now on 9
 - clang: 4 (2017-03) - now on 10
-- Qt: 5.9 LTS (2017-05) - now on 5.14
+- Qt: 5.11 (2018-05) - now on 5.14
 
 For reference, the platforms currently (2020-03-31) available in our CI. I use the MSVC version numbers since Windows
 is pretty great about backward compatibility:
--- a/rfcs/0001-make-platform-support-guarantees-explicit.md
+++ b/rfcs/0001-make-platform-support-guarantees-explicit.md
@@ -29,6 +29,17 @@ of legitimacy and forethought.
 The development community for SuperCollider has agreed on some guarantees about support for various platforms and
 toolchains. Anything that breaks these guarantees should be fixed as soon as possible.
 
+These guarantees should not be taken as an indication that we do not intend to support older platforms. On the contrary,
+care should be taken not to lose support for platforms with which SuperCollider is currently compatible, unless there
+are good reasons for doing so, and patches to add or restore support for any platform are welcome.
+
+Reasons to drop support include:
+- newer versions of tools and libraries have dropped support
+- security issues
+- the OS vendor itself has dropped support (for example, in Windows or macOS SDKs)
+- significant features require a feature not available on an older platform, and there is no practial way to make the
+  feature configurable
+
 These guarantees are given as both a number of years and number of releases. Whichever period is longer applies. For
 instance, macOS is supported for 2 years and 2 major releases. That means that any new version of macOS released in
 the last 2 years is supported, _and_ that the last two major releases of macOS are always supported. At
--- a/rfcs/0001-make-platform-support-guarantees-explicit.md
+++ b/rfcs/0001-make-platform-support-guarantees-explicit.md
@@ -158,7 +158,7 @@ would be:
 
 - Xcode: 10 (2018-06) - now on 11
 - MSVC: 2017 (2017-03) - now on 2019
-- gcc: 6 (2017-04) - now on 9
+- gcc: 7 (2017-05) - now on 9
 - clang: 4 (2017-03) - now on 10
 - Qt: 5.11 (2018-05) - now on 5.14
 
--- a/rfcs/0001-make-platform-support-guarantees-explicit.md
+++ b/rfcs/0001-make-platform-support-guarantees-explicit.md
@@ -30,9 +30,9 @@ The development community for SuperCollider has agreed on some guarantees about 
 toolchains. Anything that breaks these guarantees should be fixed as soon as possible.
 
 These guarantees are given as both a number of years and number of releases. Whichever period is longer applies. For
-instance, macOS is supported for 2 years and 2 minor releases. That means that any new minor version of macOS
-released in the last 2 years is supported, _and_ that the last two minor versions of macOS are always supported. At
-present (2020-04-02) that would be macOS 10.14 Mojave and 10.15 Catalina.
+instance, macOS is supported for 2 years and 2 major releases. That means that any new version of macOS released in
+the last 2 years is supported, _and_ that the last two major releases of macOS are always supported. At
+present (2020-04-02) that would be macOS Mojave 10.14 and Catalina 10.15.
 
 These guarantees are for the official releases published on GitHub and the supercollider.github.io website
 (pre-compiled binaries on macOS and Windows, and a built-from-source binary on Linuxes).
@@ -40,8 +40,8 @@ These guarantees are for the official releases published on GitHub and the super
 ### Platforms
 
 The following platforms are supported:
-- macOS: 2 years, 2 minor releases
-- Windows: 4 years, 4 releases
+- macOS: 2 years, 2 major releases (i.e. Mojave 10.14, Catalina 10.15)
+- Windows: 4 years, 4 major updates
 - Ubuntu: 4 years, 2 LTS releases
 - Debian: 4 years, 2 major releases
 - Fedora: latest release
@@ -139,7 +139,7 @@ None I can think of at present, everyone benefits from greater transparency and 
 For reference, the supported versions of the toolchains and platforms listed above for today's date (2020-03-31)
 would be:
 
-- macOS: 10.14 (2018-09)
+- macOS: Mojave 10.14 (2018-09)
 - Windows: Windows 10 Version 1607 (2016-08)
 - Debian/Raspbian: 9.0 Stretch (2017-06)
 - Ubuntu: Xenial 16.04 LTS (2016-04)
--- a/rfcs/0001-make-platform-support-guarantees-explicit.md
+++ b/rfcs/0001-make-platform-support-guarantees-explicit.md
@@ -1,0 +1,112 @@
+- Title: Make Platform Support Guarantees Explicit
+- Date proposed: 2019-10-06
+- RFC PR: https://github.com/snappizz/rfcs/pull/1
+
+# Summary
+
+We should decide on the kind of support we will guarantee for various platforms and compilers. These guarantees should
+be documented and enforced through continuous integration.
+
+# Motivation
+
+This was previously touched on in https://github.com/supercollider/supercollider/issues/2143, "specify minimum build
+requirements for each platform, scsynth/supernova". Much of what I'm writing/proposing here is the same.
+
+A project like SuperCollider has two opposing goals: provide support for as many systems as possible, and allow for
+efficient development and support of new versions of operating systems and compilers. At a certain point, it is not
+worthwhile for developers to expend effort supporting out-of-date systems or compilers, especially when they prevent
+new features or bug fixes from being implemented.
+
+In addition, documenting the guaranteed support of a project results in fewer suprises for users and new contributors,
+and reduces the need to continually discuss whether or not to break support when the desire arises. Having a
+documented policy arround such support allows breaking changes to happen with less controversy, by providing a sense
+of legitimacy and forethought.
+
+# Specification
+
+## Support guarantees
+
+Here is a proposed wording:
+
+> SuperCollider makes a best-effort attempt to provide support for software released in the last N years, including
+> operating systems, compilers, and other software required to build the project.
+>
+> For software released more than N years ago, no support is guaranteed, nor will any additional development effort be
+> expended to test or support such software. However, developers will not break compatibility needlessly, and will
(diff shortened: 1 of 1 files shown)


rfcs/0001-make-platform-support-guarantees-explicit.md
```diff
@@ -0,0 +1,152 @@
...
+Here is a proposed wording:
+
+> SuperCollider makes a best-effort attempt to provide support for software released in the last N years, including
+> operating systems, compilers, and other software required to build the project.
+>
+> For software released more than N years ago, no support is guaranteed, nor will any additional development effort be
+> expended to test or support such software. However, developers will not break compatibility needlessly, and will
+> attempt to maintain support if there is no additional cost for doing so.
+>
+> Support will not be broken in patch releases; only major and minor releases.
```
nhthn: as a clarification, i suggest "in patch releases of SuperCollider." also, something i'd like to throw in: "Breaking support for old platforms should be clearly advertised in the changelog."

rfcs/0001-make-platform-support-guarantees-explicit.md
```diff
@@ -0,0 +1,152 @@
...
+>
+> A "software release" is considered to be a major, non-patch version of an operating system or other software. For
+> instance, macOS 10.12 Sierra was first released September 20, 2016. SuperCollider's support guarantee for macOS
+> 10.12 Sierra extends until September 20, 20xx, and not until N years after July 19, 2017 (the first release of macOS
+> 10.12.6), nor N years after September 26, 2019 (the most recent release of macOS 10.12.6). Similarly, only major
+> releases of MSVC, LLVM CLang, and GCC, and minor releases of CMake are considered under this policy.
```
nhthn: i like the idea of adding this to the project docs. once made official, we should link it from the header of CHANGELOG.md.

one thing i'd like to see in this text is a complete list of the versioned entities that we need to be aware of. as a starting point, drawing off your Discussion section:

- macOS
- Windows
- Xcode
- gcc
- clang
- MSVC
- Qt
- possibly Boost?

by the way, you mentioned clang -- is that just for clang-format, or are people compiling SC with clang?

shimpe:
Looks good, but some thoughts:

Someone will need to keep the list up-to-date with newly appearing versions of compilers and frameworks and prune old versions. 

For old versions, a failing CI could be the trigger to discuss whether to fix the build or to remove the version, but for newer versions I think someone will need to actively track and try out the releases of all these new versions (unless there you wait for build problems to be reported? After all, "guarantees" are just that: guarantees that some combinations will work, and perhaps not an exhaustive listing of all supported combinations.)

mossheim:
Thanks for reading it over @shimpe! 

Project maintainers, including myself, will be the ones responsible for updating CI and the documentation of support. That documentation can be reviewed prior to each minor release, since it will not change during patch releases.

Typically we already get reports of build failures quite quickly on most platforms. For instance, we had issues reported about gcc 8, Boost 1.71, and Xcode 11 quite early on as they were released. But of course we can also periodically check our CI providers for new software availability. Again, immediately prior to a minor version release candidate would be an ideal time.

You bring up a good point though -- I was so focused on the "old" side of things I forgot to consider the new. Really what I am after here is a fair and documented method for phasing out support of outdated systems that we can no longer afford to support.

In any case, I think the way this should be phrased in documentation of supported platforms is "SuperCollider is tested and known to work on these platforms and software combinations."

mossheim:
Note - homebrew recently dropped support for macOS 10.12, which came out in Sept 2016, so I'd say 2-3 years of support on macOS is the best we can do.

rfcs/0001-make-platform-support-guarantees-explicit.md
```diff
@@ -0,0 +1,152 @@
...
+
+Since it is typically much easier to upgrade operating systems than software, I suggest N=3 for operating systems and
+N=2 for other software. In any case, I strongly suggest N<=4.
```
nhthn: N=3 for OS and N=2 for other software sounds reasonable to me

nhthn (commented):
great. this would definitely be valuable to the project. a few minor comments/suggestions

rfcs/0001-make-platform-support-guarantees-explicit.md
mossheim: agreed. will write more below

rfcs/0001-make-platform-support-guarantees-explicit.md
mossheim: will elaborate more below. we do not check it in CI, but people really should be able to build with clang if they want

mossheim:
updated. i am going with 2 years for all apple products and Qt, and 4 years for Windows, Linux, and non-Apple compiler toolchains.

A 4 year line seems to be still quite flexible for the systems and toolchains we would be supporting. See the notes at the end of the new document. I chose 2 years for Qt because 4 years would put us at 5.6 which we already don't support, and 3 years would put us at 5.9 which is a lot of versions in between. I'm not sure though. it looks like debian stretch only has Qt 5.7, but the Qt installer can easily give newer libraries. i would suggest saying 2 or 3 years for Qt is *guaranteed* but we support back to 5.7 for now.

more comments to come but that's all i have time for now.

muellmusik:
Hey Brian, looks sensible. One question: for OSes why not major versions rather than years? That seems to me likely to be more relevant. That said I know that Mac has been doing one a year, although some of these seemed fairly minor in terms of breaking stuff. (Not like the old days, though of course YMMV).

mossheim:
> for OSes why not major versions rather than years? That seems to me likely to be more relevant

i don't understand what you mean by "more relevant", sorry.

muellmusik:
> i don't understand what you mean by "more relevant", sorry.

Just that releases are perhaps a better measure of change than years.

joshpar:
As in - “for Apple, we support the current and last major OS release.” E.g. If Apple stays on OS X 10.15 for four years, we’ll keep supporting it and 10.14.

Right now, Apple does do a OS release every year, but maybe that will change? It did for MS...

/*
Josh Parmenter
www.realizedsound.net/josh
*/

muellmusik:
> As in - “for Apple, we support the current and last major OS release.” E.g. If Apple stays on OS X 10.15 for four years, we’ll keep supporting it and 10.14. 

Yup.

> Right now, Apple does do a OS release every year, but maybe that will change? It did for MS...

Yes, that's true. Of course the more regular releases have meant less change to some extent, so it's hard to quantify. But this at least is a crude measure of change.  If there's no major release for 2 years and nothing has changed it would seem odd to just drop off the last version.

mossheim:
don't the same issues arise if an OS vendor speeds up their release schedule? if we're suddenly nominally dropping support for versions released just 1 year or 6 months ago that doesn't seem very good either. how about "2 years or 2 major releases, whichever is greater"?

also, i am wondering about Fedora -- I see Fedora 30 and 31 are the only 'officially supported' versions of Fedora, but [fedora 30 was only released less than a year ago](https://fedoraproject.org/wiki/Releases/30/Schedule). @patrickdupuis any thoughts (you are the only fedora user i know)?

@dvzrv do you have any requests or points of clarification regarding these guarantees in relation to Arch packaging?

muellmusik:
> don't the same issues arise if an OS vendor speeds up their release schedule? if we're suddenly nominally dropping support for versions released just 1 year or 6 months ago that doesn't seem very good either. 

Fair point.

> how about "2 years or 2 major releases, whichever is greater"?

That sounds like a reasonable way to manage expectations. :-)

patrickdupuis:
Given the quick release cycle and short lifetime of releases, I don't think we should make any specific support guarantees towards Fedora. Its users are far less likely to take concern with our support guarantees anyway. You might want to ask the person (Fernando?) who packages for CCRMA.

mossheim:
> Given the quick release cycle and short lifetime of releases, I don't think we should make any specific support guarantees towards Fedora. Its users are far less likely to take concern with our support guarantees anyway. You might want to ask the person (Fernando?) who packages for CCRMA.

@patrickdupuis i just want to clarify, you're saying that we shouldn't make any effort to support Fedora at all? these guarantees will be used to decide whether to close or deprioritize tickets. i don't think we should ignore Fedora users entirely there. the bare minimum in my opinion would be to say that the latest release of SuperCollider should be compatible with and buildable on the latest release of Fedora. what do you think?

i think we will have to pay attention to how the 'patch releases keep minor release's platform support' part of this works in practice. it may not be possible to maintain that given how CI pipelines and package managers update.

mossheim:
updated, thanks @patrickdupuis and @muellmusik . linux support is now expressed per-distribution, and actually this year/release split was very useful to describe Qt support (2 years and 1 LTS release). nothing much changed except that now instead of supporting back to Fedora 25 it's only Fedora 31, the latest release.

muellmusik:
Hi Brian, this looks good. One small suggestion for clarity: I think MacOS 10 is only a _major_ release in the same sense that SuperCollider 3 is. We've had 10 for almost 20 years now. So calling 10.14 a minor release might be a bit confusing and actually doesn't fit with normal usage in the press etc. Not sure of the best rewording. You could say major release of MacOS 10? (I guess if we get 11 all bets are off anyway...)

mossheim:
this is a document for developers and so i think we should go by technically accurate terminology rather than what Apple or the press use. user-facing documentation will always give the explicit version numbers, as in the examples shown in this document, so there will be no room for misinterpretation.

muellmusik:
Well, I think that's confusing the terminology with the numbering system. This is ironic since we use the same one in SC, and our 'minor' releases can have breaking changes. But okay, as long as everyone's clear.

mossheim:
the names major.minor.patch are more widely adopted than semantic versioning, which is the standard that imparts specific meaning to increments in version number. SuperCollider has minor releases with breaking changes, but we still call them minor releases. c.f. discussions on [Wikipedia](https://en.wikipedia.org/wiki/Software_versioning#Sequence-based_identifiers), [StackExchange](https://softwareengineering.stackexchange.com/questions/166215/when-do-you-change-your-major-minor-patch-version-number), [StackExchange](https://softwareengineering.stackexchange.com/questions/3199/what-version-naming-convention-do-you-use).

now that i'm at my laptop I looked at the terminology Apple uses, and I see I was wrong. 'macOS' is actually meant to refer to the series of 10.x releases, and not the Apple desktop operating system in general. I was hoping this document could outlive 10.x, but probably Apple will either never release an 11.0 or they will simply increment 10.x indefinitely like Windows appears to be doing with Windows 10. in either case this document's wording will have to be updated. I also realized you have a good point about using press/vendor terminology; since I use vendor-specific terminology for Ubuntu, Qt, and Windows ("LTS" version, "update" vs. "release") i should do the same here. i've changed it to 'major' and added explanatory notes. how's that?

muellmusik:
Sure, I'm happy with whatever you think is best.

patrickdupuis:
@brianlheim sorry I missed your reply (not keeping myself on top of things right now)

I think what I meant was, "let's not specify a number of prior versions or number of years for Fedora support". According to this (https://fedoraproject.org/wiki/Releases), Fedora puts out approx. 2 releases a year. They End-of-life their releases very quickly also (Fedora 30 is EOL May 2020, while Fedora 32 is aiming for an April 2020 release). So, our minimum support for Fedora should definitely be "latest release".

patrickdupuis:
This is a great document @brianlheim. The only thing I have to suggest is, under the Platform section, we should add the names of the official distros we support next to RPI and BBB (Raspbian and Debian).

mossheim:
thanks @patrickdupuis @muellmusik ! updated, and also clarified that 'guaranteed support' doesn't mean we have open season to drop support on everything else. :)

mossheim:
i still want some time to figure out how this should be documented, ideas welcome! after that i think this should be ready for final comment period.

mossheim:
this RFC is now in Final Comment Period. resolution date is April 20th.

joshpar:
rad. thanks for this @brianlheim - I look forward to us documenting and supporting this RFC

mossheim:
i began work on expanding our Travis CI build matrix to cover some of these guaranteed toolchains.

platforms:
- ubuntu 16.04, 18.04
- macos 10.14 (travis does not have a 10.15 image yet)

compilers:
- xcode 10 and 11
- gcc 6 and 9 (the gcc version available on debian, and then newest version available)
- clang 4, 7, and 10 (oldest, middle, and newest version available)

additional configurations:
- gcc 9 with `-DSC_QT=OFF`
- gcc 9 with `-DSYSTEM_BOOST=ON -DSYSTEM_YAMLCPP=ON`

currently, [all but the last configuration are passing](https://travis-ci.org/github/brianlheim/supercollider/builds/672753341). that is because the latest boost version available on Ubuntu Bionic is quite old (1.65, from 2017). does anyone know of a way to obtain a newer version? i attempted to use [this PPA](https://launchpad.net/~mhier/+archive/ubuntu/libboost-latest) but it was not picked up correctly by cmake. if not, i will add a build for it on macOS instead since [the latest boost version in homebrew is 1.72](https://formulae.brew.sh/formula/boost#default).

mossheim:
i added a 'use system libs' build for macOS and [it passed on the first try](https://travis-ci.org/github/brianlheim/supercollider/jobs/672758201). @dvzrv it's not a Linux build but would this satisfy your request that we add a CI job to exercise the "SYSTEM_BOOST" and "SYSTEM_YAMLCPP" configuration options? from my perspective this is closer to want you want, since homebrew updates quite rapidly and any issues with newer boost versions will probably be visible to us unless you find them first :)

mossheim:
i added builds for gcc 7, 8 and clang 5, 6, 8, and 9, which helped me catch a few more holes in certain compatibility checks. that now makes 15 total ([passing!](https://travis-ci.org/github/brianlheim/supercollider/builds/672808001)) jobs in the matrix, and overall this still runs faster than the 2 jobs on appveyor due to parallelism restrictions and caching. i don't know if we would want to keep all 15 jobs, but i found it useful to make this check at least once.

mossheim:
final comment period has ended, this RFC is passed :) thanks all for the productive discussion!

mossheim:
as detailed above i've already started implementing the CI matrix for this; i will try to make updates to documentation this week and a PR for the CI sometime soon afterward

mossheim:
current implementation status:
- [x] wiki updates: https://github.com/supercollider/supercollider/wiki/Platform-Support
- [x] release instruction updates: https://github.com/supercollider/supercollider/wiki/git-workflow-and-guidelines#releasing
- [ ] updates to website: https://github.com/supercollider/supercollider.github.io/pull/214
- [ ] updates to readme: https://github.com/supercollider/supercollider/pull/4905
- [ ] new CI jobs: https://github.com/supercollider/supercollider/pull/4906

PR #10, opened 2020-06-15
Replace oppressive terminology with more accurate alternatives

mossheim:
https://github.com/brianlheim/rfcs/blob/topic/replace-oppressive-terminology/rfcs/0010-replace-oppressive-terminology.md

### DIFF
--- a/rfcs/0010-replace-oppressive-terminology.md
+++ b/rfcs/0010-replace-oppressive-terminology.md
@@ -15,29 +15,60 @@ governing the SuperCollider project:
 > A primary goal of our community, our user groups and mailing lists, symposia and other events is to be inclusive to the largest number of contributors, with the most varied and diverse backgrounds possible. As such, we are committed to providing a friendly, safe and welcoming environment for all.
 
 To that end, we should, when possible and practicable, replace usage of oppressive and exclusive language with inclusive
-and accurate alternatives.
+and accurate alternatives. This ensures that for a diverse set of new users, SuperCollider will make them feel welcome
+when reading the documentation and source code.
 
 A full motiviation and argumentation for this change, along with examples of precedent, can be found in the [Internet
 Engineering Task Force](https://tools.ietf.org/id/draft-knodel-terminology-00.html) RFC on the same topic. This RFC
-itself should be regarded as an effective "resubmission" of the IETF RFC for the SuperCollider community; I believe it
-makes the case very well, I agree with it completely, and I have nothing to add to it.
+itself should be regarded as an effective "resubmission" of the IETF RFC for the SuperCollider community.
 
 # Specification
-
-(Much of this is directly adapted from the IETF RFC above)
 
 We should:
 - **Change the 'master' branches of repositories 'master' to 'main'**, as shown in https://www.hanselman.com/blog/EasilyRenameYourGitDefaultBranchFromMasterToMain.aspx.
 - Prevent new commits from being pushed to 'master' branches of repositories.
-- Replace the oppressive terms “master-slave” with more accurate alternatives.
+- Keep inactive 'master' branches on repos where they currently exist, for historical purposes.
+- For new repos in the SuperCollider organization which are not being forked, name the default branch 'main' and do not
+  create a 'master' branch.
+- Replace terms using 'master' with 'main', for example 'master volume' to 'main volume'.
+- Replace the oppressive terms “master-slave” with more accurate alternatives:
+    - primary-secondary
+    - conductor-follower
+    - active-standby
+    - primary-replica
+    - writer-reader
+    - parent-helper
+    - coordinator-worker
 - Replace the oppressive terms “blacklist-whitelist” with more accurate alternatives.
+    - blocklist-allowlist
+    - block-permit
+- When referencing other documents or specifications that have not replaced oppressive technology, provide another term
+  with a note that their term is original and not being suggested by us
 - Use the neuter “they” as the singular pronoun and
 - Reflect on our use of metaphors generally.
 
 These changes should be made across all SuperCollider repositories and documentation.
 
-Luckily, these terms see very limited use within our repositories and code already; the change would not be difficult.
-And could be done by one or a few people within a few days.
+## Impact
+
+Luckily, these terms see very limited use within our repositories and code already; the change would not be difficult,
+and I would estimate it can be done by a few people working over a few days.
+
+Here is a complete list of documents in supercollider/supercollider which contain 'slave':
+- HelpSource/Classes/SyncSaw.schelp
+- examples/demonstrations/stealthissound.scd
+
+'blacklist':
+- none, but there is one occurrence in supercollider/hidapi
+
+'whitelist':
+- none
+
+For inappropriately gendered language:
+
+- HelpSource/Overviews/Operators.schelp and others refer to a function as a "poor man's gaussian".
+
+Probably other language worth changing could be found with more thorough searching.
 
 # Drawbacks
 
@@ -53,23 +84,3 @@ they may not receive updates. I expect the impact of this to be small, since:
   accordingly.
 - the most active repository in this project, supercollider/supercollider, does not even use 'master' as its default
   branch (it is 'develop').
-
-Changes to terminology in documentation have no drawbacks that I can see.
-
-# Unresolved questions
-
-Should the "master" branches of our repositories remain present even if they are no longer updated? I think this may be
-unfortunately desirable for the following reasons:
-- links to SuperCollider source code should be preserved when possible
-- removing the branch may be very disruptive to people who have cloned the project
-- removing a branch on which releases are tagged 
-
-These are the same reasons why I would also not propose to rewrite the history of this project by filtering each past
-commit to change this terminology.
-
-The guide I found for this, https://www.hanselman.com/blog/EasilyRenameYourGitDefaultBranchFromMasterToMain.aspx, also
-does not recommend entirely deleting the master branch.
-
-Personally, I feel that creating a new 'main' branch, and continuing work along it, is fully within the spirit of this
-change, but I also recognize that as someone who is not significantly affected by this terminology, I cannot and should
-not be the last word on this.
--- a/rfcs/0010-replace-oppressive-terminology.md
+++ b/rfcs/0010-replace-oppressive-terminology.md
@@ -5,7 +5,7 @@
 # Summary
 
 Replace oppressive terminology with more accurate alternatives across the SuperCollider organization. In particular,
-avoid use of "master/slave", "blacklist/whitelist", and use gender-neutral language.
+switch all 'master' branches to 'main', avoid use of "master/slave", "blacklist/whitelist", and use gender-neutral language.
 
 # Motivation
 
--- a/rfcs/0010-replace-oppressive-terminology.md
+++ b/rfcs/0010-replace-oppressive-terminology.md
@@ -47,7 +47,10 @@ We should:
 - Use the neuter “they” as the singular pronoun and
 - Reflect on our use of metaphors generally.
 
-These changes should be made across all SuperCollider repositories and documentation.
+These changes should be made across all SuperCollider organization (github.com/supercollider) repositories and documentation.
+
+These changes do not impact repositorities in the supercollider-quarks organization. If or when such changes are to be
+made, they will be addressed by a new RFC.
 
 ## Impact
 
--- a/rfcs/0010-replace-oppressive-terminology.md
+++ b/rfcs/0010-replace-oppressive-terminology.md
@@ -1,0 +1,75 @@
+- Title: Replace oppressive terminology with more accurate alternatives
+- Date proposed: 2020-06-15
+- RFC PR: https://github.com/supercollider/rfcs/pull/0010
+
+# Summary
+
+Replace oppressive terminology with more accurate alternatives across the SuperCollider organization. In particular,
+avoid use of "master/slave", "blacklist/whitelist", and use gender-neutral language.
+
+# Motivation
+
+According to the [Code of Conduct](https://github.com/supercollider/supercollider/blob/develop/CODE_OF_CONDUCT.md)
+governing the SuperCollider project:
+
+> A primary goal of our community, our user groups and mailing lists, symposia and other events is to be inclusive to the largest number of contributors, with the most varied and diverse backgrounds possible. As such, we are committed to providing a friendly, safe and welcoming environment for all.
+
+To that end, we should, when possible and practicable, replace usage of oppressive and exclusive language with inclusive
+and accurate alternatives.
+
+A full motiviation and argumentation for this change, along with examples of precedent, can be found in the [Internet
+Engineering Task Force](https://tools.ietf.org/id/draft-knodel-terminology-00.html) RFC on the same topic. This RFC
+itself should be regarded as an effective "resubmission" of the IETF RFC for the SuperCollider community; I believe it
+makes the case very well, I agree with it completely, and I have nothing to add to it.
+
+# Specification
+
+(Much of this is directly adapted from the IETF RFC above)
+
+We should:
+- **Change the 'master' branches of repositories 'master' to 'main'**, as shown in https://www.hanselman.com/blog/EasilyRenameYourGitDefaultBranchFromMasterToMain.aspx.
+- Prevent new commits from being pushed to 'master' branches of repositories.
+- Replace the oppressive terms “master-slave” with more accurate alternatives.
+- Replace the oppressive terms “blacklist-whitelist” with more accurate alternatives.
+- Use the neuter “they” as the singular pronoun and
+- Reflect on our use of metaphors generally.
+
+These changes should be made across all SuperCollider repositories and documentation.
+
+Luckily, these terms see very limited use within our repositories and code already; the change would not be difficult.
+And could be done by one or a few people within a few days.
+
+# Drawbacks
+
+Aside from the humanpower required to make such changes, very few:
+- People will have to know to switch from 'master' to 'main' , and this may lead to a small amount of confusion. Since
+  most (in fact, I think all) repositories within SuperCollider do not commit directly to the default branch, and
+instead work by opening PRs against whatever the default branch happens to be, much of this change will be entirely
+transparent to contributors and require no effort on their part.
+
+One case where this may lead to confusion is that if people continue to attempt to pull from 'master' but not 'main',
+they may not receive updates. I expect the impact of this to be small, since:
+- the change will be advertised, and so most people who have already cloned SuperCollider will know about it and update
+  accordingly.
+- the most active repository in this project, supercollider/supercollider, does not even use 'master' as its default
+  branch (it is 'develop').
+
+Changes to terminology in documentation have no drawbacks that I can see.
+
+# Unresolved questions
+
+Should the "master" branches of our repositories remain present even if they are no longer updated? I think this may be
+unfortunately desirable for the following reasons:
+- links to SuperCollider source code should be preserved when possible
+- removing the branch may be very disruptive to people who have cloned the project
+- removing a branch on which releases are tagged 
+
+These are the same reasons why I would also not propose to rewrite the history of this project by filtering each past
+commit to change this terminology.
+
+The guide I found for this, https://www.hanselman.com/blog/EasilyRenameYourGitDefaultBranchFromMasterToMain.aspx, also
+does not recommend entirely deleting the master branch.
+
+Personally, I feel that creating a new 'main' branch, and continuing work along it, is fully within the spirit of this
+change, but I also recognize that as someone who is not significantly affected by this terminology, I cannot and should
+not be the last word on this.


Spacechild1:
> Replace the oppressive terms “master-slave” with more accurate alternatives.

I don't get it... "master" is not an oppressive term in itself. The metaphor "master/slave" might be seen as problematic by some people (especially in the US), but we don't use the word "slave" in the context of git anywhere...  I always thought of the master branch as the master copy, like a "vinyl master".

To be honest, renaming the master branch feels like virtue signalling to me. I'm all for creating inclusive environments, but this is a bit ridiculous. I don't want to start a flame war, just giving my honest opinion :-).

If you decide to go with this, please keep master as an alias, so you don't break existing scripts.

> Replace the oppressive terms “blacklist-whitelist” with more accurate alternatives.

Agreed.

> Use the neuter “they” as the singular pronoun and

Agreed.

> Reflect on our use of metaphors generally.

Agreed.

patrickdupuis:
This post from the GNOME dev mailing list from 2019 offers an explanation of the origin of git`s `master` branch: https://mail.gnome.org/archives/desktop-devel-list/2019-May/msg00066.html

It appears that git inherited `master` from Bitkeeper which names repositories (not branches) master and slave.

Spacechild1:
I know. But is this really how the term "master" is actually perceived in the context of git? I certainly didn't know it came from master/slave, I just found out today, and so did most people I dare say. Again, git doesn't use the word "slave" anywhere.

jamshark70:
Tbh, when I was working on my DDW___Clock classes, it did cross my mind, "Why do we call them master and slave? Is that really *good*?" But I ignored that moment of unease and went with it because it's standard terminology.

So, movements to change standard terminology do make a difference, and I'll go change my quark at some point.

On the question of the master branch -- apart from breaking scripts, it hurts no one to call it "main" or even "stable," whether or not there's a current explicit reference to slavery. I do see Christof's point -- I had never thought of other git branches as slaves -- but I don't see a compelling reason to keep the current term.

I'm not crazy about bandwagoning, but in this case, I did have misgivings about the words in my own work, so I appreciate the nudge.

Edit: Of gender-neutral pronouns, I quite often like to use "she" especially for fields that are traditionally male dominated. It can give male chauvinists an extra moment of cognitive dissonance that they might not feel with "they" and "them."

clfest:
This is extremely silly, and has a touch of Newspeak in it. The meaning of a word is context driven. You don't get rid of 'bad thoughts' by banning words that are in some way associated with them.

joshpar:
I support this change. The underlying context of our language on a daily basis informs our and our community's world view. If we make an effort to direct our language towards an inclusive example, that helps move society as a whole in that direction over time.
@clfest - Newspeak is saying one thing, and meaning another. If the purpose was to change to Main but still with the intention of keeping a master / slave relationship, I would agree there is a problem. I think this is a move away from that.

eleses:
Frankly "master" for a/the git branch is probably inappropriate even on a purely technical level as there's no real sense in which the other branches are "slaves". They're all just refs as far as I understand git... (which is not incredibly well, I admit.)

As for the other uses in SC tree... I'm surprised they are that many in (just in c/cpp/h/hpp/sc/schelp):

```
Search "master" (126 hits in 35 files)
```

A bunch of those come from boost and portaudio actually... (The latter uses "master volume" a lot, and so does the wasapi.) But then so does `Server.schelp`.

Since the IETF [draft](https://tools.ietf.org/html/draft-knodel-terminology-01) doesn't seem to discuss what's the best replacement for "master volume" (only "master" in some other contexts), does anyone know what proposed terminology alternatives are in the music/sound world? So that if we replace "master volume" with something less oppressive it would not be a term completely unique to SC...

N.B. the IETF draft proposes these alternatives to "master-slave":

 o  Primary-secondary
 o  Leader-follower
 o  Active-standby
 o  Primary-replica
 o  Writer-reader
 o  Coordinator-worker
 o  Parent-helper

I guess "primary volume" would be most suitable in a sound context... Opinions?

Surprisingly, I only got one hit for "slave" in my search; that's in SyncSaw.schelp.

It looks like we don't have to do any changes for "blacklist" as there are no occurrences in SC proper (well, there's one in portaudio's source) and likewise for "whitelist".

Also, as far the order of changes goes, I'd suggest starting with the help files as probably the most visible to SC users. The (SC) code internals can be done on a lower priory "pass".

lnihlen:
While some may not be bothered by master/slave references it's their privilege to not be bothered by that. It costs the dev team literally nothing to adjust our language to be more inclusive. Huge supporter of these practices, and will adjust Scintillator and Hadron structure to suit.

jamshark70:
> The latter uses "master volume" a lot, and so does the wasapi.

Another way to look at it is that the objectionable concept is that of a slave. "Master" as a word has multiple connotations, some of which imply subjugation of human beings (or autonomous agents), others of which do not.

A year or two ago, I was working on networked clock classes, and I did call one the "master clock" and the other, the "slave." That's a direct reference to the idea of one agent subjugating another, and I did have an icky feeling about that (though I didn't act on it at the time -- now I find myself glad that someone is pushing for a change).

To me, a master volume or master channel is quite different. Nothing is enslaved; the source channels have their own automation and effects processing, and they continue working individually. (Yes, it's true that if you turn the master/global volume all the way down, then you hear nothing, but all of the sources are still processing as if the volume were up.)

The intention behind changing the terminology is good. It's also possible to dilute the good effect by misapplying the sensitivity where it may not make sense (such as the incident a few years ago, where a US Representative was called out for using a word that only sounds like an objectionable slur but whose meaning and etymology have exactly nothing to do with the Latin root for "black" -- the objection was well-meaning but misplaced).

eleses:
@jamshark70  Well, if master by itself is not objectionable, then there isn't much work to do.  "slave" only appears in one SC file...

But my understanding of the gnome discussion (linked in a [prior comment](https://github.com/supercollider/rfcs/pull/0010#issuecomment-644492527) in this thread)  is that "master" in itself is objectionable, even with no mention of "slave".

See also, github as a whole making this change https://www.zdnet.com/article/github-to-replace-master-with-alternative-term-to-avoid-slavery-references/ And discussion in the git developer community itself https://lore.kernel.org/git/CAOAHyQwyXC1Z3v7BZAC+Bq6JBaM7FvBenA-1fcqeDV==apdWDg@mail.gmail.com/ (linked from that piece of news)

Interesting technical discussion from there, regarding `git` proper:

```
From: Junio C Hamano <gitster@pobox.com>
Date: Sun, 14 Jun 2020 14:06:18 -0700

Konstantin Ryabitsev <konstantin@linuxfoundation.org> writes:

> - having a branch named "master" is already not required, so any 
>   existing software that expects there to always be a "master" branch is 
>   already broken and wouldn't get any more broken by the move away 
>   towards more descriptive terminology

The above is mostly true but not entirely, I have to remind you.

There certainly is the concept of the primary branch in each
repository.  With the current version of Git, it is hardcoded to be
the 'master' branch, and we are going to make it configurable with a
configuration variable [*1*].  There are a few examples that the
primary branch is treated specially:

 - When merging into the primary branch, the auto-generated merge
   message is different from a merge into any other branches.  This
   was because most of the merges, especially in early days of Git,
   were into the primary branch (there weren't many cross-branch
   merges made) and we did not want to see repeated "Merge X into
   'master'", "Merge Y into 'master'", etc.---we just say "Merge X",
   "Merge Y", etc. when merging into the primary branch.

 - "gitk" gives preferencial treatment to 'master' when there are
   too many references it has to choose which ones to make visible.

 - "git fast-export --anonymize" redacts the name of all the
   branches, but the name of the 'master' branch is left intact in
   its anonymized output (which is done in order to make the primary
   line of work identifiable even in the redacted output stream).

There may be others.  Software that assumes that 'master' is special
is *not* broken as you stated (we will break them when we allow
changing the default---that is the cost those of us who are working
on the transition plan thought worth paying).  The concept of "there
is one thing that is special among others" can serve useful purpose.

It of course opens a different can of worms ;-) Even though we can
please master-slave-haters by moving away from the particular word
'master', those who cannot stand the very idea of one thing being
special among others will not be satisfied (and we shouldn't even
try to please them, IMO).



[Footnote]

*1* There is a related but separate concept of the "default name"
for the primary branch in newly created repositories, which also is
hardcoded to be 'master'.  We are going to make it configurable,
too.  This controls the name used by "git init" (possibly we will
also add a command line override "git init -b name" to countermand
the configured default) and "git clone" (which tries to use the
name of the branch pointed at by HEAD of the other side but has to
fall back to something when it cannot figure it out).
```

muellmusik:
Do it. The point of inclusive practice is not just about whether something offends us as individuals, but whether it might offend others.

Yes master has multiple meanings, but as has been pointed out for years, its use in sound work is often pretty weird metaphorically, and possibly sexist. (Master tape? Surely it should be mother tape?) Good metaphors also aid understanding.

jamshark70:
> ... as has been pointed out for years, its use in sound work is often pretty weird metaphorically, and possibly sexist.

That sounds compelling to me. Though I'll note that this discussion appears at the same time that Peter Kirn writes about the topic in CDM and says essentially the same as I did (which I hadn't noticed on first skimming) -- "The term master on its own you don’t have to worry about. It’s associated with mastery, master classes, master copies, master bedrooms, the antiquated practice of calling young men “masters,” Masters’ degrees, uh… MasterCard. Not one of those terms originated in the practice of slavery. (I checked.)" https://cdm.link/2020/06/lets-dump-master-slave-terms/

Yes to evaluating metaphors on their merits (in particular, thumbs up to "mother tape" but I don't expect specialists to start rebranding themselves as "mothering engineers" anytime soon). "On their merits" may end up concluding that a specific use of a specific word has nothing to do with an objectionable origin, however. ("Master" is admittedly troublesome as the oldest meanings of the word do connote one person with power over another, though. It may not be possible to recuperate that word in every case.)

eleses:
And I think the notion of "master volume" has lot more to do with a notion of [slave (amplifier)](https://www.google.com/search?&q=slave++amplifier+master+volume), than the "master" git branch ever had to do with any notion of slave...

Windows get around this issue by calling it "device volume" https://superuser.com/questions/1202337/how-can-i-sync-all-sources-in-windows-volume-mixer

But in SC that's not quite correct. So maybe just "server volume"?

muellmusik:
Andra McCartney was writing on this stuff from a feminist perspective already decades ago.

https://muse.jhu.edu/article/585329/summary

Spacechild1:
@jamshark70 thanks for your nuanced comments! This is the discussion I was hoping for (instead of people just silently downvoting other's comments *cough*).

For everyone: I'm not *against* renaming master to main, I just don't see the point. For me, this is just creating unnecessary friction for no apparant gain. But if you decide to do so, please keep it as an alias, so you don't break existing scripts.

On a side note, GitHub advertizing this change feels more like a PR stunt. If they want to do something meaningful, they should officially stop cooperating with ICE... Just my two cents.

Spacechild1:
> N.B. the IETF draft proposes these alternatives to "master-slave":
>
> ...
> o Leader-follower
> ...

Now this is funny. As an Austrian I find this metaphor troubling because of our own history. This just shows how much the whole discussion depends on the cultural context. I just think that proposing/imposing a change in language is a very radical act that needs to be considered carefully.

miguel-negrao:
I agree with the general intention of this RFC. In general I agree that language always evolves and we can change it if we think that is warranted. 

I agree with avoiding use of master/slave, that has such a specific meaning, there is no way to avoid the original meaning.

In the case of _parts_ of words which don't have a specific oppressive meaning, but only gain that meaning depending on the context and history of use of the term (such as terms with black or white), I think it would be prudent to investigate the etymology of the word, and also consider if it has been used with an oppressive meaning. 

Regarding whitelist and blacklist, I found [this article/commentary](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC6148600/) which  suggests that it has a racist origin, so If that is the case I would agree with avoiding it.

Nevertheless it would make language poorer if no words including black and white could be used when also having a specific positive or negative association. For instance, from a quick googling, "whitewash" which has a negative association might come from painting a wall white (which is a common color), therefore rendering invisible the imperfections and cracks that existed before. We should just make sure the words do indeed have a racist connotation, which is not the case of all words containing black and white which have a positive or negative connotation.

Regarding the use of singular they pronoun, had a [a quick look at what that is](https://en.wikipedia.org/wiki/Singular_they), and to a non-native english speaker it feels extremely confusing.  I don't have an issue with people using it here, but I don't think it should be expected that non-native english speakers should be capable of using it. We have something similar in Romance languages for formal treatment (third person singular or second person plural) and non-native speakers really struggle with that.

eleses:
Regarding pronouns: most of the SC documentation I've looked at is written in 2nd person "you can..., you should..." etc. And the things being referred to in 3rd person are inanimate, so "it", "its" etc. Thus, I don't see a concerted effort being needed on that pronouns angle, but of course any slip-ups should be corrected when found. 

(If SC had documentation e.g. in French or some other language with [gendered nouns](https://en.wikipedia.org/wiki/Grammatical_gender#Mostly_semantic_criteria), it would be potentially more work in regard.)

telephon:
Btw. the term "master" in the context of a repository may well come from production, in particular printing, like of vinyl and CD, where you have a master from which the copies are made. The German word is beautiful: "Druckvorlage". I like this shift of semantics, because we can think of a "master clock" in the same way. That means, good replacements for master are synonyms in that sense: model, prototype, source.

Now that the whole issue is about a _pair_ of terms, the replacement needs to be paired as well. I'm not a native speaker, so I may miss out a good solution. One reasonable possibilty is `source` vs. `sink`, another one `sender` and `receiver`.

I personally always found that "server - client" has a pretty bad taste, but others may not be similarly inclined?

(As we are at it, and try to become more politically aware, we may also move our repository to gitlab, as to avoid the master silo we are entering ... But this is another issue altogether)

cianoc:
Blacklist and whitelist aren't racist. Their use comes from England and their origins are in middle English. 

I agree with James that replacing the use of 'he'/'his' with 'she'/'hers' etc is a good idea. It reads much better than 'they', is less confusing for non-native speakers and at least makes the documentation more inclusive.

I can see an argument for changing master to main for clarity. I think it makes it harder to understand the decentralized nature of git. As far as oppressive language goes...  ¯\_(ツ)_/¯ 

Serious question - what terminology would you use for a MIDI clock slave?

Spacechild1:
Just to give some more perspectives, here's what some POC developers have to say:

https://news.ycombinator.com/item?id=23505713
https://news.ycombinator.com/item?id=23501861
http://antirez.com/news/122#comment-4084872912

It turns out that some find the GitHub initiative patronizing and insulting... Now what do you do? I just want to point out that things are not as clear cut as some people might believe.

miguel-negrao:
> Serious question - what terminology would you use for a MIDI clock slave?

Parent/child,  Writer/reader ?

muellmusik:
> Serious question - what terminology would you use for a MIDI clock slave?

In Utopia there is a simple decentralised clock which uses Conductor/Follower (I think this was Alberto's suggestion). A conductor coordinates, but it's ultimately a collaborative arrangement! (Infamously dictatorial maestros notwithstanding.)

Spacechild1:
I think parent/child is not too bad. It emphasizes the hierarchical structure of a tree of devices.

cdbzb:
the MIDI clock *must* follow - its not collaborative and that's what the term 'slave' captures. I'd suggest the terms 'driving/driven' (from gears) for that one... so 'driving-clock' 'driven-clock'

cdbzb:
'whitelist' and 'blacklist' are nice English words - it would be nice to replace them with similar words... what about redlist and greenlist like traffic lights?

cdbzb:
master/copy could be source/clone perhaps

cianoc:
Do black people complain about the usage of master/slave? I could give you a long list of local black complaints, and while there are symbolic fights (the confederate flag, removing confederate statues), language only comes up when it's used in a deliberately offensive way. For example there was a local politician who used to refer to black men as 'boy'.

lnihlen:
Why is it the responsibility of black folks to determine if terminology is racist or exclusive? I find the terminology offensive. We need to do better. It is our responsibility to make our community as inclusive as possible. We should not just wait until somebody is hurt by this language.

Spacechild1:
> Why is it the responsibility of black folks to determine if terminology is racist or exclusive?

Because it's about them?

>  I find the terminology offensive. 

And some black software devs find the initiative to rename the master branch insulting and patronizing, see https://github.com/supercollider/rfcs/pull/10#issuecomment-644791473

So who wins?

> We should not just wait until somebody is hurt by this language.

Of course we should. Proactively banning language is ridiculous IMO.

cdbzb:
Nonsense. It's about us.

We use the term 'force quit' but not 'rape quit'. Why? Because we don't
need to be making that metaphor part of our creative work. If slavery is a
shame and a disgrace, why choose that metaphor to have to read and write
while we try to make music? Driver clock and driven clock communicates the
relationship between MIDI clocks more clearly without bringing unpleasant
human relationships to mind, so why not

As far as the linked posts, it's not either/or. We can make these small
aesthetic changes and the more important political changes as well. These
tiny changes can serve as small reminders in this small but, to me,
precious space

So full steam ahead, change it up.

Spacechild1:
> Nonsense. It's about us.

I don't think it's our right to decide for a black person what is racist or not.

> We use the term 'force quit' but not 'rape quit'. 

That is something entirely different. Rape culture affects all people, so it's also about "us".

> If slavery is a shame and a disgrace, why choose that metaphor to have to read and write
while we try to make music?

Actually, I've agreed that the master/slave metaphor might be problematic and that we can come up with a better one, see https://github.com/supercollider/rfcs/pull/10#issuecomment-644802457.

But the "master" branch is *not* a reference to slavery. Changing it just because it might eventually offend someone just doesn't make any sense to me...

lnihlen:
> > Nonsense. It's about us.
> 
> I don't think it's our right to decide for a black person what is racist or not.

Racism impact everyone. Not just black folk. White people are responsible for racism and need to take action. 

> 
> > We use the term 'force quit' but not 'rape quit'.
> 
> That is something entirely different. Rape culture affects all people, so it's also about "us".
> 
> > If slavery is a shame and a disgrace, why choose that metaphor to have to read and write
> > while we try to make music?
> 
> Actually, I've agreed that the master/slave metaphor might be problematic and that we can come up with a better one, see [#10 (comment)](https://github.com/supercollider/rfcs/pull/10#issuecomment-644802457).
> 
> But the "master" branch is _not_ a reference to slavery. Changing it just because it might eventually offend someone just doesn't make any sense to me...

It offends me. So we're not in the realm of the theoretical here. I am offended.

Spacechild1:
@lnihlen May I ask you *why* you're offended? I'm honestly trying to understand it.

lnihlen:
No you may not.

Spacechild1:
Ok, I respect that.

cdbzb:
@christof  - you don't know my background - nor that of most of the posters
here. Why not just step off?

mossheim:
hi all,

first off, thank you everyone for your participation in this conversation. i understand this is not an easy topic, and discussing it publicly -- and in a place where, among others, current and potential employers and professional contacts may be watching -- is sure to add an extra level of stress for many.

i am unfortunately not able to respond to every point that has been raised thus fair in this reply. i would like to, but i still have my own thinking and reading to do. but, there are a few things that i want to note now:

---

one. this RFC is not a referendum on a sociopolitical movement or anyone's perspective on it. it is a very concrete proposal, referencing very concrete and specific arguments. please take the time to read the proposal -- including the IETF RFC -- and discussion thus far before commenting. it is extremely disrespectful to everyone here to offer your opinion in passing on what you *guess* this proposal is about and the reasons why it has been made.

---

two. regarding "keep master as an alias of main for scripts". i am strongly opposed to adding such a policy to this RFC. my reasoning is:
- many tools and scripts that interact with GitHub repositories simply use the 'default' branch of the repository and do not specify a branch. this change would be transparent to them. in some ways, this could even be considered good practice.
- i do not recall any major issues when the default branch of supercollider/supercollider became `develop` instead of `master` several years ago, so i am skeptical that this change would be disruptive in any way
- the primary goal of this proposal is to migrate away from language which is objectionable and inaccurate. continuing to support scripts and/or documentation which use this terminology would undermine that goal

---

three. regarding "allow using she as a gender-inclusive pronoun." i strongly disagree with adding such a policy to this RFC. my reasoning is:
- language choices should be oriented toward promoting inclusivity. whether our language forces men and/or male chauvinists to reckon with their discomfort is strictly a secondary concern.
- using "she" instead of "he" still excludes and marginalized nonbinary people. out of "he", "she", and "they", it is only "they" which allows all readers to insert themselves into a passage without misgendering themselves.
- many well-regarded style guides already recommend 'they', among them [the Purdue Online Writing Lab](https://owl.purdue.edu/owl/general_writing/grammar/pronouns/gendered_pronouns_and_singular_they.html) and [APA Style](https://apastyle.apa.org/style-grammar-guidelines/grammar/singular-they). More examples can be found on [Wikipedia](https://en.wikipedia.org/wiki/Singular_they#Acceptability_and_prescriptive_guidance). i have yet to find a style guide which regards 'she' as a gender-inclusive pronoun

---

four. regarding "do not require nonnative speakers of English to use gender inclusive language (specifically 'they'/'them'/'theirs')." i strongly disagree with adding such a policy to this RFC. my reasoning is:
- being welcoming to nonnative speakers and nonbinary people are non mutually exclusive goals. and of course, there are many nonnative speakers of English who are also nonbinary people
- we can be polite in correcting one another on language. presumably, a hypothetical nonnative speaker of English also shares in our common goal of creating an inviting and inclusive space, and can understand the need for using gender-inclusive language. if they cannot, this has less to do with being a non-native speaker of English, and more to do with holding negative and/or marginalizing attitudes about trans and nonbinary people.
- we already expect everyone involved in this community to use respectful and inclusive language regardless of their background. to say that nonnative speakers should not be required to use or learn inclusive language in this single instance raises the question of why nonbinary people in particular are deserving of such a compromise.
- nonnative speakers are able to learn inclusive language; however, trans and nonbinary people are not able to learn to not be marginalized by exclusive language and misgendering

---

i would like to add that i consulted with a nonbinary colleague in developing the above reasoning; i owe several of these arguments to them.

other points which i note, but do not have time at present to address:
- "use etymology as a determining factor in deciding whether language is inclusive or exclusive"
- "whitelist and blacklist are not oppressive" (variety of reasons given)
- "the word 'master' in itself is not a reference to slavery, unless it is accompanied by the term 'slave', and therefore things named 'master' should not necessarily be renamed"
- "nobody is actually being harmed by this language"

other points which i note, and will not address because they have no bearing on the proposal at hand:
- "this is virtue signalling"
- "people are too sensitive"

EDIT -- i have made my best effort to represent the viewpoints above in my summary. i do this to help myself and others keep track of the conversation, which has been thankfully so active and thoughtful already. if you feel i have misrepresented something you said, please contact me privately or leave a comment here, my goal is not to belittle anyone with my paraphrasing.

mossheim:
@cdbzb @Spacechild1 @lnihlen 

out of respect for everyone who is taking the time to discuss this very nuanced and complex topic, might i kindly suggest taking a bit more time to think before you respond? we owe it to each other to keep this discussion focused and substantive, and i am a bit worried by the personal nature of the last few replies. i don't want to silence good discussion where it is happening, but i also want to make sure this conversation remains organized and respectful.

mossheim:
i would also like to offer the opinion that when discussing a topic as sensitive as this, using the emoji reactions in the github UI hinders good faith participation and encourages polarization and competitiveness. it reminds me of why i left social media platforms long ago and avoid them to this day. i will not go so far as to say i don't want people to use them here, but please take that into consideration as we continue the discussion.

lvm:
@brianlheim +1 to the proposed changes. Thanks for bringing this up.

Unfortunately I don't have terms to suggest as replacements (plus as a non-native-english-speaker i might miss some nuances) but I'd like to think  that whatever is decided will probably be better than what we currently have.

Spacechild1:
@brianlheim I just want to say that I respect your RFC. I don't agree in all points (as you've probably found out), but I won't oppose it. (I feel I don't even have the right to do so, since I'm only a very casual contributor.)

Your answer regarding "keep master as an alias of main for scripts" is convincing. You're right that scripts shouldn't assume that the default branch is "master", since this is just a convention but not mandated by git. So yes, the change shouldn't be too disruptive and the objections are mainly political.

Are you planning to discuss this further in a developer meeting? I would like to take the chance and talk about this a bit "in person" (if I'm welcome ;-). If you feel that's not necessary, just tell me honestly.

patrickdupuis:
@Spacechild1 developer meetings have been less frequent over the last few months (for obvious reasons). There are currently no meetings planned. Meetings and meeting polls are announced on scsynth.org. the Slack, and on the mailing list. Everyone is welcome to join.

cianoc:
I think MIDI slave is an example of where you might need to be very careful as that's the technical term widely used in the industry. So changing the term in help files could confuse people.

Sciss:
Thanks Brian. I think _main_ is a good name for the default branch. Following the general motion in combing through software, I looked through my own code and found it straight forward to replace something like _master volume_ by _main volume_ as well.

In my SC UGen database, I found one occurrence in `SyncSaw`, in which I now use _synchronizing (principal) frequency_ for the former "master" oscillator frequency, and _subordinate sawtooth frequency_ for the former "slave" oscillator frequency.

__Edit:__ I second the use of the _they_ pronoun. I think it's basically standard now in English, and I find it very easy to use in my own texts. The fact that other languages like German lack such a concise solution is no argument against it.

muellmusik:
> the MIDI clock _must_ follow - its not collaborative and that's what the term 'slave' captures.

Um. Performers who don't follow the conductor are not collaborating (okay Zorn game pieces aside). The point is that it's a question of collaborative roles. One coordinates, but that doesn't have to mean they're the 'master'.

muellmusik:
> 'they'/'them'/'theirs'

In the parallel discussion on the users list, @miguel-negrao helpfully linked to the wikipedia article on singular _they_ in English. It's gone in and out of fashion, true, but has been part of the language since the 14th century. It's very much a part of the language, and I don't think we need to question its validity in any way. https://en.wikipedia.org/wiki/Singular_they

miguel-negrao:
> four. regarding "do not require nonnative speakers of English to use gender inclusive language (specifically 'they'/'them'/'theirs')." i strongly disagree with adding such a 
policy to this RFC. my reasoning is:

As a disclaimer, I find being welcoming to nonbinary people an important goal (thank you @brianlheim for calling attention to that), I just wonder what is the best way to do that. I don't have an issue with using gender-neutral language, I was raising some questions about the obligatory use of 'they'.

Let me be more specific about what I find will be quite hard for non-native speakers to start using, regarding 'they'. I find it confusing to conjugate in the plural when referring to a single person. Googling a bit more it seems this is also confusing to at least some native english speakers. From the wikipedia page:

"They are my child."

The verb indicates a plural, multiple people, and then what comes after indicates a single person. Learning to write these type of sentences requires re-learning the patterns that one is used to considering "wrong". Isn't this sentence wrong according to the english grammar being taught today in schools ?

This probably also creates ambiguous sentences regarding the plural and singular:

"When are they going to merge this PR ?"

Is "they" a team ? Is it a single person ?

I wonder if there are alternatives for avoiding writing he or she which don't require writing this type of phrase with mixed singular/plural ? 

In this specific case one could write:

"When is <person's name> going to merge merge this PR ?"

[Wikipedia](https://en.wikipedia.org/wiki/Non-binary_gender) also mentions some gender-neutral pronouns (ze, sie, hir, co, and ey) which don't require changing the way verbs are conjugated. 

What do you think of changing

- Use the neuter “they” as the singular pronoun and

to

- Use gender-neutral language. Do not use the pronouns "he" and "she". The use of the neuter “they” as the singular pronoun is recommended.

?

This leaves some space to use other alternatives while being gender-neutral.

jamshark70:
It's often the case that documentation and/or technical advice can be written in plural form: instead of "the user should configure this setting in his/her startup file," it could be "users may configure this setting in their startup file." Then there is no grammatical question *and* we get gender neutrality.

I also want to say thanks to Brian for bringing up nonbinary people -- I should have considered that but didn't.

swannodette:
Hullo as a relative newcomer to the SuperCollider project, I respect the tone and reasoned practicality of this RFC, thank you @brianlheim. It looks good to me. 

I'm a person of color and to be honest I've been in the technology game long enough that the ubiquity of master/slave and gendered documentation in engineering, software, and music technology has become taken for granted. I'm jaded to the point I don't really notice it much anymore. However, I don't think such changes are even about users who have already been ground down by the ubiquity of problematic technological terminology. Open source efforts are only as alive as their community - one can make changes that either invest in the continued growth of a community or contribute to its eventual demise. If SuperCollider is to be an essential creative tool for a diverse set of *new* users, the RFC ensures that these people will feel welcome when reading the documentation and source code.

FWIW, I feel like Moog has tried to set a better example these days. Looking over their online videos, one sees a more balanced representation of musicians using their gear. Scanning through the Moog Grandmother manual there's exactly one use of the term Master - "Master volume". The section on oscillator sync completely drops using any master/slave terminology and opts for a equally clear explanation that doesn't sound awkward in the slightest.

If Moog can do this, but a sophisticated electronic music tool like SuperCollider cannot, it does not speak very highly of the project's leadership nor its community.

This seems like a minor change with little impact for development that ensures that young musicians of all backgrounds looking for a new exciting creative tool won't be turned away by terminology that's clearly on the way out elsewhere.

mossheim:
thanks all for another day of respectful and constructive discussion on this RFC! responses to specific points are below.

---

regarding "leader/follower" as an alternative to "master/slave": i think there are enough alternatives that we can simply choose to strike this one from the list of suggested replacements. @Spacechild1 is not the only person from Europe to take issue with this term. Florian Grond on the mailing list [wrote](https://www.listarc.bham.ac.uk/lists/sc-users/msg67834.html):

> coming from Europe I am always surprised by what positive connotations words like leader or leadership have in other parts of the world.

in the next draft of my RFC, i will note this.

---

regarding @Spacechild1 's request to discuss this in a sc-dev meeting: i personally would not have the energy and mental stamina to lead a real-time conversation about this. i already spend about 3-4 hours a day right now thinking about everything that has been said and writing my responses. @joshpar has offered to lead that discussion though. since the conversation itself centers around language which many including myself find offensive and/or marginalizing, i would suggest that we talk about it at the end of the meeting so that anyone who would like to avoid it can still feel comfortable joining the call. sound good @Spacechild1 ?

---

regarding @miguel-negrao 's comment about singular "they" and confusion resulting from it use. i think @jamshark70 makes a good point that often gender inclusivity and clarity in documentation can easily be achieved together. i think there has perhaps been some misunderstanding about this line in the IETF RFC. what i would suggest as wording for this RFC, based on consultation with other style guides including the AP and Purdue OWL guides, is as follows:
- use the singular "they" when referring to a person of unknown gender expression, including a hypothetical person
- when possible, craft technical writing to avoid ambiguities resulting from the use of singular 'they'/'them'/'theirs'

(of course, once we know a person's preferred pronouns, we should use those. i am not including that statement in this RFC because our community code of conduct already states that we should be respectful in our communication.)

the fact of the matter is that there are already people out there right now who use "they" as their preferred pronoun. in order to be inclusive of these people, we all (including native speakers) may need to relearn our patterns of thinking, writing, and speaking to accommodate this, regardless of the difficulty we may have in doing so.

recognizing that grammar is just one aspect of culture, this reminds me of the discomfort some people have felt learning that they can no longer ask me, a ~~non-straight man~~, "do you have a girlfriend?". growing up, they were taught in school and in books that boys have girlfriends, and girls have boyfriends, and it was difficult for them to recognize that they live in a world where this no longer applies. i even remember a very similar concern about the use of the gender-neutral term "partner" -- won't there be some confusion, about whether it means romantic partner or business partner? jokes about this ambiguity even made rounds in sitcoms at the time. but in the end, there are very few cases where the ambiguity actually results in genuine confusion, it can almost always be gathered from context. all those jokes did, really, was reinforce the notion for me and other non-straight people that our sexuality was a burden to others. @miguel-negrao , i understand you mean no offense, and i do understand your concern comes from a desire to understand and to be respectful. however, i would like to offer the opinion that the best way forward is to use the terms people already say they prefer to start, give ourselves some time to get used to them in practice, and decide later whether there is more to consider.

to sum up, many style guides, including the ones i linked above, have agreed that singular "they", and not any other pronoun or set of pronouns, ought to be used when referring to a person of unknown gender, and i will not consider alternatives to that recommendation for this RFC.

---

regarding the points in my last comment that i said i note, but "have no time at present to address", i have decided not to respond to them, except to say that i do not agree with them, and i think they have already been addressed by others.

---

> This seems like a minor change with little impact for development that ensures that young musicians of all backgrounds looking for a new exciting creative tool won't be turned away by terminology that's clearly on the way out elsewhere.

this is a really excellent way of phrasing it, and i feel the same. thank you @swannodette !

---

other points that i don't have time to address at the moment, but do note for later:

- @cianoc 's note about the term "MIDI slave", and how to replace terminology in such away that we avoid reinforcing oppressive terms while not harming the clarity of our documentation. i think we do need some specific guidance on that and it shouldn't be hard to find that by looking at other parallel conversations and proposals to this one.

miguel-negrao:
@brianlheim thank you for your comments. 

>  i think there has perhaps been some misunderstanding about this line in the IETF RFC. what i would suggest as wording for this RFC, based on consultation with other style guides including the AP and Purdue OWL guides, is as follows:

The wording you suggest seems good to me, thank you. I would suggest adding a first line about using gender-neutral language in general, to make it clear what is the general goal and to make sure all different issues with using non-gender-neutral language are covered.

>  @miguel-negrao , i understand you mean no offense, and i do understand your concern comes from a desire to understand and to be respectful. however, i would like to offer the opinion that the best way forward is to use the terms people already say they prefer to start, give ourselves some time to get used to them in practice, and decide later whether there is more to consider.

Agreed, this sounds like a good suggestion. Perhaps a link could be placed somewhere to a good reference for how to use the singular they, for those that are not familiar.

As an aside, I read a bit more about the singular they, and I found personally that some uses of it already feel as natural, or even more natural to me, than the grammatically correct version, probably because they have been in use for a while already. I'm thinking of sentences such as:

"One of your friends phoned, but they didn't leave a message."
"No one put their hand up."
"The person you mentioned, are they coming?"
"Everyone doubts themselves at one time or another."

They all seem to have in common referring to unknown people. I suppose the sentences with singular they which personally I will need more getting used to are the ones which refer to a known person (in those cases I would probably avoid the pronoun by writing the person's name).

Just to make sure it's clear, I'm in favor of approving this RFC, and I thank @brianlheim for introducing it.

ps: I really like allowlist/blocklist, the meaning is just transparent.

Spacechild1:
> regarding @Spacechild1 's request to discuss this in a sc-dev meeting: i personally would not have the energy and mental stamina to lead a real-time conversation about this. i already spend about 3-4 hours a day right now thinking about everything that has been said and writing my responses.

@brianlheim I totally understand. These kind of heated topics can drain a lot of energy. And I didn't mean it as a "request", rather as an invitation :-).

> i would suggest that we talk about it at the end of the meeting so that anyone who would like to avoid it can still feel comfortable joining the call. sound good @Spacechild1 ?

This sounds good to me. As I've said, I won't oppose the RFC (it has no practical ramifications for me), but if someone wants to learn the broader reasoning behind my objections (and those of other people coming from a similar stance), I would be happy to explain and discuss it in a friendly private video conversation where we can all talk freely while seeing each other's faces :-). But it is not essential.

mossheim:
@miguel-negrao @Spacechild1 great! thank you both for understanding.

---

a few times now it has been pointed out that there are additional terms that others might find disagreeable. two such examples:
- https://github.com/supercollider/rfcs/pull/10#issuecomment-644767771: "I personally always found that "server - client" has a pretty bad taste, but others may not be similarly inclined?"
- https://www.listarc.bham.ac.uk/lists/sc-users/msg67875.html: "I am blind.  I would appreciate it if people would stop bringing up my disability as a negative in casual discussions. However, I am fully aware that my wish will never be granted, because there is no movement to educate the public regarding my disability and how and when to use it in common speech."

i would like to note a few things:

---

one. the [diversity statement](https://github.com/supercollider/supercollider/blob/develop/CODE_OF_CONDUCT.md#diversity-statement) in our code of conduct says:

> We encourage everyone to participate and are committed to building a community for all. Although we will fail at times, we seek to treat everyone both as fairly and equally as possible. Whenever a participant has made a mistake, we expect them to take responsibility for it. If someone has been harmed or offended, it is our responsibility to listen carefully and respectfully, and do our best to right the wrong.
>
> Although this list cannot be exhaustive, we explicitly honor diversity in age, gender, gender identity or expression, culture, ethnicity, language, national origin, political beliefs, profession, race, colour, immigration status, education level, size, family status, religion, sexual orientation, class, socioeconomic status, and technical ability. We will not tolerate discrimination based on any of the protected characteristics above, including discrimination against participants with disabilities.
>
> We are aware that the definitions and examples for diversity and discrimination have a strong cultural bias and are likely to change over time. They mirror the sensitivities of the SuperCollider community at the time of writing. We welcome suggestions as to how to extend our understanding of discrimination and exclusivism.

regardless of the passing or rejection of this RFC, if someone already feels that others in SuperCollider spaces will not or are not able to respect their wishes around language that harms them, **this is our failure as a community**. we should reflect on this and do better, especially those of us in moderation positions such as myself.

---

two. i would like to clearly state that this RFC is not intended as a valuation of some people's dignity above others'. i apologize if it has appeared that way. i chose to focus on the terms i did because:
- the IETF RFC was very well argued and seemed very reasonable to me
- i do not consider myself expert or well-read enough on these topics to offer additional terms
- there are no terms i wanted to add which i find personally insulting or oppressive toward me, and therefore felt it better to not attempt to speak "on behalf" of any other group

---

three. at the same time, i have decided that, despite the comments i quoted above, i will not be adding any additional terms for replacement to this RFC. this is because:
- this discussion has already generated 50+ comments in a few days, and therefore i believe it is better to keep further discussion constrained to the terms at hand
- if this RFC is passed, it should be seen as community agreement that terms which are offensive to others, and which we can easily replace with terms that are at least as clear and as technically accurate as the originals, ought to be replaced
- in fact, it is my hope (perhaps a foolish one) that this is the **only time** an RFC will be required to make such replacements, and that future requests for terminology changes, building on this discussion as a precedent, can be agreed upon and implemented far more easily

---

four. one of the points of action in the linked IETF RFC is:

> reflect on [our] use of metaphors generally

i would like this point to please not be overlooked; it is an integral part of this proposal.

Spacechild1:
@brianlheim After thinking some more about the whole issue, I want to provide the following honest thoughts, which hopefully are perceived as constructive and help to understand my view point:

---

Personally, I would appreciate it if the RFC would not simply assert that the words in question *are* oppressive. The meaning of words or metaphors always depends on the social and cultural context and I think the RCF should reflect that. Several people on the mailing list and in this thread have expressed doubts or concerns about changing certain terminology. (I'm not talking about those who just said "this is silly", but those who brought actual arguments.)

I'm not asking to address each of those concerns, I think some kind of general acknowledgement would be sufficient. This could be as simple as replacing the sentence

> To that end, we should, when possible and practicable, replace usage of oppressive and exclusive language with inclusive and accurate alternatives.

with something in the line of

> To that end, we should, when possible and practicable, replace usage of terminology that some people might experience as oppressive and exclusive with more neutral and inclusive alternatives. We are aware that whether a given terminology is perceived as problematic or not is to a large extent determined by the socio-historic and cultural context, but when in doubt we opt for maximal inclusiveness.

This would help to make the RFC more nuanced and less US-centric and acknowledge cultural differences, while keeping the original intention.

---

Some general remarks:

Millions of people (including myself) are practically forced to learn and write/speak English as a second language, because it is the de-facto lingua franca. (Don't worry, I love the language nevertheless :-) Countless varieties of English are spoken all over the world, the majority being "non-native" speakers. To put it bluntly: the USA shouldn't have a monopoly on the meaning of English words. In practice, the USA *has* a huge global influence - which creates a power relationship that we all should be aware of.

I want to give an example: For many people the word "master" is almost entirely free of any notions of slavery (let alone racism) and has largely positive connotations. Suddenly being told that the word "master" is in itself oppressive/offending can lead to confusion and might even result in strong emotional responses (as we've seen). The better approach would be to explain that the word "master" has certain negative connotations in the USA (or elsewhere) and then *ask* other countries to reconsider their usage or understanding of the word out of mutual respect.

ghost:
> White people are responsible for racism and need to take action.

This is plain nonsense !

mossheim:
hi all,

i have just updated the RFC text. i encourage everyone to look at the diff and read the new document. a descriptive overview of the changes:
- remove 'unresolved questions' section on deleting master branches; i have decided to propose that we not delete master branches entirely, but discontinue committing to them
- flesh out 'specification' section with a list of suggested replacements drawn from the IETF RFC and from this thread
- respond to @cianoc's comment about how we will refer to terminology in other specifications or documents
- add "Impact" section with discussion of the scope of expected changes

i now feel like this RFC is close to being in its final form. however, i am going to wait a bit longer before starting a final comment period since there has been a lot of discussion and the changes in this last commit are substantive.

---

@Spacechild1 thank you for your comment, it certainly helps me to understand your perspective on this in a way that your earlier comments did not. however, i have decided not to change the wording of my RFC according to your suggestion because i feel that it would obscure the clarity of the original.

rfcs/0010-replace-oppressive-terminology.md
```diff
@@ -15,29 +15,60 @@ governing the SuperCollider project:
...
+- Replace the oppressive terms “master-slave” with more accurate alternatives:
+    - primary-secondary
+    - conductor-follower
+    - active-standby
+    - primary-replica
+    - writer-reader
+    - parent-helper
+    - coordinator-worker
 - Replace the oppressive terms “blacklist-whitelist” with more accurate alternatives.
+    - blocklist-allowlist
```
telephon, line 43: blocklist-passlist ?

cdbzb:
redlist greenlist?

mossheim:
@telephon @cdbzb 

thanks for the suggestions. i am leaving it as blocklist-allowlist for now, because this is the terminology proposed in the linked IETF RFC and because it is already being adopted elsewhere, notably in [chromium](https://beebom.com/google-blacklist-whitelist-removed-code/).

in the future can you please offer a brief rationale for suggested changes? this will help us avoid bikeshedding and keep discussion focused.

telephon:
> in the future can you please offer a brief rationale for suggested changes?

Sorry, could you explain? I don't understand what you mean by this …

mossheim:
@telephon you made a suggestion yesterday without explaining why, for instance, you thought it would be an improvement, or whether it should augment or replace the existing text. please do that for future suggestions.

madskjeldgaard:
I support this RFC. Good work @brianlheim !

telephon:
> @telephon you made a suggestion yesterday without explaining why, for instance, you thought it would be an improvement, or whether it should augment or replace the existing text. please do that for future suggestions.

ok, I see what you mean. I suppose this is true not just for me ;)

brunoruviaro:
I just read the proposal in its entirety, and I wholeheartedly agree with the changes proposed there. Thank you.

mossheim:
thanks again everyone for the thoughtful and respectful conversation.

i now feel ready to put this RFC into [Final Comment Period](https://github.com/supercollider/rfcs#finalizing-an-rfc). the date to merge or reject this RFC is **July 12**.

eleses:
Can we have some decision on "master volume"? As I mentioned previously it's by far the most common use of the "master" term that I found in SC itself. (The other things are nice to have been decided in theory, but their practical impact in terms of changes required will be close to zero, except for the git/github branch name issue.)

mossheim:
@eleses Please, read the RFC before commenting. I have already covered your question in the text:

> - Replace terms using 'master' with 'main', for example 'master volume' to 'main volume'.

eleses:
@brianlheim I missed the [update](https://github.com/supercollider/rfcs/pull/10/commits/ab006d3cd383fcc1fd7ac9ca537296444a21f7c3#diff-b226b83dd9595e8418e59ab1fe69bc99R33), sorry.

mossheim:
no worries! I do appreciate that you wanted to follow up on that. :)

spluta:
I have read and I fully support these changes. Thank you for the opportunity given to discuss this as a community.

rfcs/0010-replace-oppressive-terminology.md
telephon: as you asked for a rationale (not next time, but this time), here it is: 

"blocking" and "passing" are of the same semantic category (physical action), just as "forbidding" and "allowing" would be (moral/legal), but not "blocking" and "permitting". This is why I looked for something that 'rhymes' better.

The same applies to block-permit, of course.

mossheim: thank you for the rationale! I am still keeping it to these terms for the reasons I gave below. at least in this point in time, a quick internet search shows that momentum is behind blocklist/allowlist as replacements.

I don't agree with the semantic argument either. these semantic categories you gave don't match with my understanding of these terms. "block" is used quite often when, for example, a book or law or legal order is prevented from taking effect. it has a clear meaning in legal contexts. to give a recent example from the news -- ["The European Union is expected to block most US residents from visiting the region when travel restarts due to rising case numbers. The EU is finalising the list of visitors allowed into the bloc as early as July. "](https://www.news.com.au/world/americans-unlikely-to-be-let-into-eu/video/b537c9abc557c66fb8b393384c69a2cc). note the pairing "block from" / "allow into".

davidgranstrom:
I agree with the RFC and support the proposed changes.

rfcs/0010-replace-oppressive-terminology.md
telephon: OK, I do see the point, although the alternative seems still clearer. 

In any case, in the linux kernel it is now:
```
+Recommended replacements for 'blacklist/whitelist' are:
+    'denylist / allowlist'
+    'blocklist / passlist'
```

https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=49decddd39e5f6132ccd7d9fdc3d7c470b0061bb

patrickdupuis:
I also have read the RFC and the linked IETF document. I fully support this RFC and the proposed changes.

adcxyz:
Hi, 
What are the plans for smoothly moving all quarks from "master" to "main" branches? 

Currently the Quark and Git classes have "master" hardcoded in 3 places, 
and I am wondering how that migration can work: 
older SC versions will keep trying to checkout "master" branches;
I assume newer ones will likely look for "main" branches?
If so, are there any smart git ways of aliasing main and master branches 
to each other for a transition period?

mossheim:
@adcxyz good question! i hadn't planned on making any changes to repositories in supercollider-quarks, to be frank because the thought just hadn't crossed my mind. that's my mistake, as RFC author the responsibility for not considering that lies with me. my interpretation is that under this RFC there would be no such migration; i was thinking here only of repositories within `github.com/supercollider`. but, i would certainly understand if other people feel that this interpretation doesn't match with what they agreed to!

i have updated this RFC to more clearly present my interpretation.

my current feeling is that, as this RFC is mergeable and has already reached a state of agreement, we should merge this, make the changes as detailed here in supercollider org repos, and leave the problem of supercollider-quarks to a future discussion.

i say this for the following reasons:
- changes in supercollider org repositories can be implemented independently of supercollider-quarks
- i suspect some people here also did not consider supercollider-quarks, since there are big technical and human questions to answer there and nobody else seems to have raised or considered them in this discussion, and so trying to decide this problem now, after people have already had their say, risks exhausting discussion participants and even appearing to change the scope while people aren't paying attention
- this brings an entirely new set of technical complications that i am not prepared to address, and i do not want to hold up these more easily achieved changes with that discussion
- supercollider-quarks are fundamentally 3rd-party code, which brings an extra dimension to this problem that i am also not prepared to address

i would like to hear from a few other people. to clarify, my proposal is that we merge this, begin to make changes in the supercollider org **only**, and at the same time launch a discussion about whether a new RFC is needed to make changes to supercollider-quarks. if people disagree with this, can you please indicate what you think should happen to this RFC instead?

please do not merge this RFC, in the meanwhile.

mossheim:
to be extra clear, as i see it the options here are:
1. consider this RFC to not address the question regarding supercollider-quarks, and merge it.
2. withdraw this RFC and submit a new one that addresses that question, so that we can attempt to reach a consensus on implications for both supercollider and supercollider-quarks at once.

i will *not* be emending this RFC to try to address this question, and i will *not* be seeking a new round of consensus/input on this RFC. it is too late in the process for me to feel comfortable doing that.

Spacechild1:
I don't want to start a new discussion, but

> Currently the Quark and Git classes have "master" hardcoded in 3 places,

My question is: why is "master" hardcoded in the first place? I think Brian's answer regarding the fear of breaking scripts also applies here:

> many tools and scripts that interact with GitHub repositories simply use the 'default' branch of the repository and do not specify a branch. this change would be transparent to them. in some ways, this could even be considered good practice.

So why not simply pull the default branch?

jamshark70:
> So why not simply pull the default branch?

That seems entirely reasonable. Individual quark authors would be able to set the branch name on their own. (Unless there's some technical reason why that wouldn't work?)

mossheim:
hey @jamshark70 @Spacechild1 -- i would really really prefer to just focus on the single question at hand, it's urgent for the future of this RFC and supersedes any other conversation related to it. also, FYI this has been discussed before [here](https://github.com/supercollider/supercollider/issues/4321) (and it appears i'm the reason why this is the case!).

jamshark70:
Alberto raises the valid point that we can't purge all occurrences of "master" from the SC repository without altering the behavior of quarks. Specifically, if a quark's default branch is not master but the author intends "master" to be the stable branch that users generally install, the choices are: a/ incompletely implement this RFC and keep backward compatibility in this case by keeping "master" in the quarks code; b/ change it and require quark authors using this configuration to update.

It doesn't affect me as I'm not separating development and stable branches in my quarks, and I'm not opposed to either a/ or b/. But I think Alberto is correct that it isn't possible to avoid all decision about this. The decision may well be a/, and I think that would be legitimate and not deeply offensive, and allow significant movement forward. It's worth making that an explicit decision (and commitment to come back to the quarks question later).

patrickdupuis:
I think what @brianlheim is proposing in this RFC only applies to repositories under the `supercollider` organization. `supercollider-quarks` is a separate organization not in scope of this RFC.

It would be best if we can keep the discussion about Quarks for another time.

mossheim:
alright! since this RFC has broad support in the form where it at least applies to the repositories in the SuperCollider org, and since we've already reached the end of the final comment period, i'll consider it passed with the caveat that more discussion/planning is needed before touching anything in supercollider-quarks. that already gives us quite a bit to implement, and i think is really the core of what this RFC was about to begin with. there are several important ongoing discussions right now on other parts of the project that i'd like to participate in, and we also have a release immanent, and my first RFC still hasn't been fully implemented yet!, so i would like to let those things settle before implementing this.

thank you again to everyone for the engaging and respectful discussion, this has made me very happy to be a part of this project!

muellmusik:
Thank you @brianlheim for taking this initiative!

smoge:
> Replace the oppressive term “master-slave” with more accurate alternatives.

 [Poor Hegel](https://www.jstor.org/stable/1344332)
 

It is clear to admit that identity politics, which seeks to promote a new type of sensitivity, contributes sufficiently to ensure a new kind of censorship. The “inclusion” and “diversity are constantly used to justify seeking freedom. As their identities refer to “symbolic” indications – or which show that they are ascending middle-class people –its lute emphasis on the use of words, the study of pronouns, imaginative etymologies, and finally, a folk bullfight of “linguistic delicacies”, which, on days of delirium, one might try to include among supposedly “racist” insults expressed as “caixa-preta” or “buraco negro”.It is worrying that criticism of identity politics and defense against censorship (even if "for the best") is associated with the Right. 

But people forget SOMETHING FUNDAMENTAL: When people think saying something changes reality, THIS isn’t true. **Actions, not just words, are what bring change.** 

I'd like to suggest an interesting thinker, maybe some of you might enjoy: https://www.youtube.com/watch?v=EQ5gLuk06TU

smoge:
The issue is when people think saying something changes reality, which isn't true. Actions are what bring change, not just words.

See another perspective: https://www.youtube.com/watch?v=EQ5gLuk06TU

PR #16, opened 2021-03-28
Improve SuperCollider versioning for non-release builds

dyfer:
https://github.com/dyfer/rfcs/blob/scversion-update/rfcs/0016-scversion-improvements.md

### DIFF
--- a/rfcs/0000-scversion-improvements.md
+++ b/rfcs/0000-scversion-improvements.md
@@ -1,0 +1,68 @@
+- Title: Improve SuperCollider versioning for non-release builds
+- Date proposed:
+- RFC PR: https://github.com/dyfer/rfcs/pull/0000
+- SuperCollider issue: 
+
+# Summary
+
+I'd like to propose introducing finer granularity in SuperCollider versioning.
+
+# Motivation
+
+Currently, SC builds other than the release builds report the version of the last released build. This pertains to the "latest develop" builds uploaded to S3, builds from other branches and PRs, as well as local builds. I would like these builds to report a more accurate version.
+
+# Specification
+
+My proposal pertains to SuperCollider version, as defined in the `SCVersion.txt` file and reported by sclang when starting up. 
+
+I am seeking feedback on two "levels of granularity" for the proposed changes:
+1) coarse, where individual builds do not get individual version suffixes, aimed at differentiating released builds from non-released builds
+2) (optional) fine, where individual builds are differentiated, in terms of version suffix
+
+For 1) I am seeking to fine-tune the naming of the general version suffixes, like `alpha`, `beta`, or `prerelease`. For 2), I'm seeking feedback whether such granularity is practical and desired, as well as the format for the version suffix.
+
+### Coarse versioning
+
+I'd like to propose to append suffixes `-alpha` for the builds on the develop branch, and `-beta` for the builds on the release branch (`3.12` etc), for the versions _following_ the last bugfix release and _preceding_ the _release candidate_ of the next minor/major release. 
+
+I propose that the builds in the `develop` branch have the version number of the _next_ release (`current minor` + 1), with the appropriate suffix.
+
+I propose that the builds in the `3.x` branch have the version number of the _upcoming_ release, with the appropriate suffix.
+
+Currently used suffixes (`-rc1` etc) would be used the same way, as they are used currently.
+
+Here's a rudimentary visualization of proposed changes, starting at the current development cycle (after the `3.11.2` release, before creating the `3.12` branch):
+```
+[branch name]
+
+| 3.12.0-alpha     | 3.13.0-alpha                                      | 3.14.0-alpha
+[develop] ________ _________________________________________ ... _____ ____________________ ...
+                  \                         [master]  __ /            \
+            [3.12] \________________________________/           [3.13] \___________________ ...
+                    | 3.12.0-beta   | 3.12.0-rc1      | 3.12.0          | 3.13.0-beta 
+```
+
+### Fine versioning
+
+Optionally, I would like to consider, whether the non-release builds could have an additional version suffix. The motivation for this is to differentiate between multiple development versions of SC that might be present on the user machine. This can happen for example, if different versions have different crucial fixes that the user needs, which have been merged into the develop branch at different points in time.
+
+I was considering adding information like short SHA or short SHA and latest commit date, in order to have a reference to both commit build, as well as a sense of _when in the development cycle the build was created_, for example:
+
+- `3.12.0-alpha+SHA`, e.g. `3.12.0-alpha+abcdef0`
+- `3.12.0-alpha+YYYYMMDD`, e.g. `3.12.0-alpha+20210401`
+- `3.12.0-alpha+YYYYMMDD.SHA`, e.g. `3.12.0-alpha+20210401.abcdef0`
+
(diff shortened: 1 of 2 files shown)


mossheim:
Thanks for this @dyfer !

> I am seeking feedback on two "levels of granularity" for the proposed changes:
>    1. coarse, where individual builds do not get individual version suffixes, aimed at differentiating released builds from non-released builds
>    2. (optional) fine, where individual builds are differentiated, in terms of version suffix

My feedback is that I mainly want (1), because in my view it is a serious problem in the way the version number is communicated, and I'm going to mostly focus on that in my comments.

I would not suggest to use "alpha" for the in-development version names. I would use "dev" instead, like we have done in the past. The reason is that "alpha" is a real identifier we might use for release names, and since there is potentially a long time between minor releases, I would only put it in the version name when we want to indicate that we are actually going to release something under that name. To me, it's clearer to say that dev is work prior to releases, and alpha and beta are releases in some sense. "Prerelease" is not good as a name here in my opinion because people sometimes say both alpha and beta releases are "prereleases", including github's own UI:

image

An alternative word I found with a quick internet search is "pre-alpha". I don't have a strong preference between pre-alpha and dev but personally I find dev less jargony.

Other than that, I think what you've proposed is good -- I am fine with using "3.x.y-beta" on release branches, since the intent of the release branch is that the software is stable and we do not add anything which would introduce new bugs. This also reduces the number of times we have to change the version file.

For (2), I do have one small question about the motivation:

> This can happen for example, if different versions have different crucial fixes that the user needs, which have been merged into the develop branch at different points in time.

I don't understand this. Has this happened? If so, is there anything else we can do to fix this situation and make sure it doesn't happen again? It seems like "a user needs to have multiple versions of SC on the machine to be productive" is a problem worth digging into more.

----

Other feedback:

> I imagine the coarse versioning to be implemented by creating two version entries in SCVersion.txt

I didn't understand this; I was imagining we would accomplish this by just updating SCVersion.txt more often than we currently do, and that this would mostly be a change in workflow/documentation. Are you suggesting something else?

dyfer:
Thanks for the feedback @brianlheim !

I'd be happy to go with `-dev` for builds from develop and `-beta` for builds from `3.x` branch.

>> This can happen for example, if different versions have different crucial fixes that the user needs, which have been merged into the develop branch at different points in time.

> I don't understand this. Has this happened? If so, is there anything else we can do to fix this situation and make sure it doesn't happen again? It seems like "a user needs to have multiple versions of SC on the machine to be productive" is a problem worth digging into more.

Not sure what you mean by "has this happened". Anyone can make that happen by downloading multiple versions of SC on their machine :) I certainly have more than one SC version on mine (under macOS). 

In the past I needed to run my own build with fixes not in the develop branch for a project, and I had the "stable" sc version as well. I think they both had the same version number, which this RFC would address. Then, when doing dev SC work I'd have other builds floating around. In that case it would be beneficial to be able to differentiate where the builds were made in the release cycle somehow (after unpacking the archive there's no way to tell a difference, aside from the build branch in the "about" dialog). But again, maybe this is not a big deal.

>> I imagine the coarse versioning to be implemented by creating two version entries in SCVersion.txt

> I didn't understand this; I was imagining we would accomplish this by just updating SCVersion.txt more often than we currently do, and that this would mostly be a change in workflow/documentation. Are you suggesting something else?

In my RFC I propose that the builds from the develop branch have a "future" version number, comparing to those from the `3.x` branch (using your suggested suffix here):
```
[branch name]

| 3.12.0-dev       | 3.13.0-dev                  
[develop] ________ ___________
                  \                         
            [3.12] \_____________
                    | 3.12.0-beta   
```
In that case we'd constantly have a merge conflicts if we just changed the version as it is implemented now. I have an idea how to address it but first I want to agree on the change itself, and then discuss implementation later.

mossheim:
> Not sure what you mean by "has this happened". Anyone can make that happen by downloading multiple versions of SC on their machine :) I certainly have more than one SC version on mine (under macOS).

OK, I think I understand now, thanks for explaining. I was under the impression some users *need* to have multiple versions of SC to get work done at all.

> In the past I needed to run my own build with fixes not in the develop branch for a project, and I had the "stable" sc version as well. I think they both had the same version number, which this RFC would address. Then, when doing dev SC work I'd have other builds floating around. In that case it would be beneficial to be able to differentiate where the builds were made in the release cycle somehow (after unpacking the archive there's no way to tell a difference, aside from the build branch in the "about" dialog). But again, maybe this is not a big deal.

This is starting to feel to me like a possible [XY problem](https://meta.stackexchange.com/questions/66377/what-is-the-xy-problem). Can we take a step back and first understand why it is that you often have multiple development versions of SC on your computer and in places where you cannot easily distinguish them? A typical workflow is to use source control to manage the different things you are working on, and rebuild the application as needed. Sometimes one may have different build directories or configurations for release and debug builds, but that distinction is made clear from the directory names of the builds.

> In that case we'd constantly have a merge conflicts if we just changed the version as it is implemented now. I have an idea how to address it but first I want to agree on the change itself, and then discuss implementation later.

This is not true; you only have to resolve a merge conflict the first time you merge the branch in. Afterward, there is no conflict. So there would only be a conflict when the release branch's version is updated, which is in my opinion not enough to justify anything more complex than what we have now.

dyfer:
> This is starting to feel to me like a possible [XY problem](https://meta.stackexchange.com/questions/66377/what-is-the-xy-problem). Can we take a step back and first understand why it is that you often have multiple development versions of SC on your computer and in places where you cannot easily distinguish them? 

Maybe it is a XY-problem. But just because it _is_ possible to have multiple versions of SC coexisting, and one might need them for compatibility with various projects, IMO it's not a scenario to be dismissed. Though probably not something to get into a lengthy discussion in this RFC.

> A typical workflow is to use source control to manage the different things you are working on, and rebuild the application as needed. Sometimes one may have different build directories or configurations for release and debug builds, but that distinction is made clear from the directory names of the builds.

I think it's not the best to assume this is the only workflow possible. As I said, one might need to have multiple SC versions when a certain version (either own build, or an old version, for compatibility) is present in order to run a certain piece/project. Should we open a separate issue to discuss this?

>
> > In that case we'd constantly have a merge conflicts if we just changed the version as it is implemented now. I have an idea how to address it but first I want to agree on the change itself, and then discuss implementation later.
> 
> This is not true; you only have to resolve a merge conflict the first time you merge the branch in. Afterward, there is no conflict. So there would only be a conflict when the release branch's version is updated, which is in my opinion not enough to justify anything more complex than what we have now.

Is that so?

I'm imagining SCVersion in develop branch would look something like this:
```
set(SC_VERSION_MAJOR 3)
set(SC_VERSION_MINOR 13) # note minor version
set(SC_VERSION_PATCH 0)
set(SC_VERSION_TWEAK "-dev")
```
while the one in 3.12:
```
set(SC_VERSION_MAJOR 3)
set(SC_VERSION_MINOR 12) # note minor version
set(SC_VERSION_PATCH 0)
set(SC_VERSION_TWEAK "-beta")
```
If the conflict needs to be resolved only once (and the version in the dev branch is maintained), that's great! I didn't think it would be that easy.

mossheim:
My suggestion at this point is that you withdraw the second part of this RFC and just keep the "coarse versioning" bit, both because it's the most important piece in my view and because the rest seems like it could benefit from further discussion and design in a way that the RFC process isn't designed to support. Feel free to make an issue ticket for it if you'd like.

dyfer:
At the dev meeting on 2021-09-05 we tentatively agreed on a part of this RFC: version on develop branch will be called one version higher than the current release branch, with an appropriate suffix (`-prerelease` or `-preview`). The details will be  further discussed in the PR suggesting that change in the main SC repo.

The more elaborate versioning connected with commit sha etc will not be implemented for the time being.

